PR #11, opened 2019-06-05
Add support for SwiftUI.Image

samuel-mellert:
I added an extension for SwiftUI.Image, comments and suggestions are welcome.

### DIFF
--- a/Sources/SFSafeSymbols/ImageExtension.swift
+++ b/Sources/SFSafeSymbols/ImageExtension.swift
@@ -1,0 +1,14 @@
+import SwiftUI
+
+@available(iOS 13.0, *)
+public extension Image {
+
+    /// Creates a instance of `Image` with a `UIImage` object that has its rendering mode set to `.alwaysTemplate`.
+    ///
+    /// - systemSymbol: The `SFSymbol` describing this image.
+    @available(iOS 13.0, *)
+    init(systemSymbol: SFSymbol) {
+        self.init(uiImage: systemSymbol.toImage.withRenderingMode(.alwaysTemplate))
+    }
+
+}
--- a/README.md
+++ b/README.md
@@ -102,6 +102,12 @@ let image2 = SFSymbol.eCircleFill.toImage(withConfiguration: /* Some UIImage.Con
 let image3 = SFSymbol._11CircleFill.toImage(compatibleWith: /* Some UITraitCollection */)
 ```
 
+... or create a `SwiftUI.Image` by initializing it using the `SFSymbol`...
+
+```swift
+Image(systemSymbol: .cCircle)
+```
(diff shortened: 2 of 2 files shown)


Sources/SFSafeSymbols/ImageExtension.swift
```diff
@@ -0,0 +1,14 @@
...
+
+@available(iOS 13.0, *)
+public extension Image {
+
+    /// Creates a instance of `Image` with a `UIImage` object that has its rendering mode set to `.alwaysTemplate`.
+    ///
+    /// - systemSymbol: The `SFSymbol` describing this image.
+    @available(iOS 13.0, *)
+    init(systemSymbol: SFSymbol) {
+        self.init(uiImage: systemSymbol.toImage.withRenderingMode(.alwaysTemplate))
```
fredpi, line 11: Just wondering, whether we should use the `.alwaysTemplate` option everywhere... or make it accessible as an optional param? However, probably 100 % of the time, we want a template image, do you agree? If yes, would you mind implementing this everywhere? ;)

Sources/SFSafeSymbols/ImageExtension.swift
```diff
@@ -0,0 +1,14 @@
+import SwiftUI
+
+@available(iOS 13.0, *)
+public extension Image {
+
+    /// Creates a instance of `Image` with a `UIImage` object that has its rendering mode set to `.alwaysTemplate`.
+    ///
+    /// - systemSymbol: The `SFSymbol` describing this image.
+    @available(iOS 13.0, *)
+    init(systemSymbol: SFSymbol) {
```
fredpi, line 10: Is there a specific reason there aren't initializers for `UIImage.Configuration` and `UITraitCollection`? If no, would you mind adding those too (Also to the README)?

fredpi (changes requested):
Thanks for the PR 🚀 I made some suggestions, would you please have a look at them? ;)

Also, after having added this, we should restructure the tests a bit, but that should be in a separate PR as it would also affect the existing tests

Sources/SFSafeSymbols/ImageExtension.swift
fredpi: Just had a look at the WWDC session about SF Symbols and they said all symbols are already rendered as `.alwaysTemplate`, so we won't need this ;)

samuel-mellert: When I tried in in Xcode with preview it didn't work, until I set the rendering mode to `.alwaysTemplate`. I'm watching the session right now. Let me get back to you afterwards.

samuel-mellert:
Reworked and created new pull request.

PR #13, opened 2019-06-05
Swift 5.1-specific improvements

knothed:
Because this is a Swift 5.1-repository (as stated in the README), this pull request establishes a Swift 5.1 feature: Returning from single-line methods without using the `return` statement.


fredpi:
Looks good! Thanks for the PR 👍

PR #18, opened 2019-06-12
Add a changelog

knothed:


fredpi:
Thanks for the PR 👍

A few notes:
• I don‘t think we should list changes of the first release ;)
• Please add a section „Unreleased“ on top with the corresponding subtitles (Added, Changed...).
• We should reference issues and authors in the changelog
• We should mention the changelog in the CONTRIBUTING.md

knothed:
@fredpi What should "Unreleased" mean? Should this section be updated each time stable is updated? Does this make sense?

knothed:
Also, I think it makes sense to outline what the capabilities of the first release were. Especially as they are currently the key part of the project.

PR #19, opened 2019-06-12
Add iOS 13-availability check to the test classes

knothed:


fredpi:
Please let the test fail, if not tested on iOS 13+

knothed:
Done.

PR #22, opened 2019-06-17
Add support for SwiftUI.Image

samuel-mellert:


Sources/SFSafeSymbols/SFSymbolExtension.swift
```diff
@@ -2,23 +2,28 @@ import UIKit
...
     @available(iOS 13.0, *)
     var toImage: UIImage {
         UIImage(systemSymbol: self)
     }
 
-    /// Retrieves the corresponding UIImage with the given configuration.
+    /// Retrieves the corresponding `UIImage` with the given configuration.
     ///
-    /// - withConfiguration: The UIImage.Configuration applied to this system image.
+    /// - Parameter withConfiguration: The `UIImage.Configuration` applied to this system image.
```
fredpi: Would you mind using the param name instead of the argument label here and at some other places to follow the same style everywhere? ;)

Tests/SFSafeSymbolsTests/ImageExtensionTests.swift
```diff
@@ -0,0 +1,19 @@
+@testable import SFSafeSymbols
+import XCTest
+
+#if canImport(SwiftUI)
```
fredpi: If we check for this, we should also wrap the entire test into the `canImport` statement and let the test fail if `SwiftUI` cannot be imported. This way we get a test failure instead of a compiler error.

fredpi:
Thanks for the PR 👍

PR #25, opened 2019-06-19
[README.md] Fix some naming, spacing and language issues

Jeehut:
Note that I started this PR mainly because I felt like the "Swift Package Manager: compatible" badge was simply too big, so using "SwiftPM: compatible" (which is the official abbreviation AFAIK) would be better. Also since there are really many package managers supported now, I felt like a separate line for those badges would also make sense, so I added a breakline. The other stuff are just small details I noticed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,9 @@
     <a href="https://github.com/piknotech/SFSafeSymbols/blob/stable/LICENSE.md">
         <img src="https://img.shields.io/badge/license-MIT-lightgrey.svg" alt="License: MIT">
     </a>
+    <br />
     <a href="https://github.com/apple/swift-package-manager">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,11 +71,11 @@ After specifying `"SFSafeSymbols"` as a dependency of the target in which you wa
 
 ### Accio
 
-Do the same configurations as for Swift PM, then run `accio update` instead of `swift package update`.
+Do the same configurations as for SwiftPM, then run `accio update` instead of `swift package update`.
 
 ### Carthage
 
-Make the following entry in your Cartfile:
+Make the following entry to your Cartfile:
```
fredpi: To unify things, let's also change "Make" to "Add"

fredpi (changes requested):
Thanks for the PR 👍 Just one little, picky thing to resolve

README.md
Jeehut: Fixed. :)

Issue #24, opened 2019-06-18
Fix deployment target


fredpi:
To import the framework with iOS versions lower than iOS 13, we'll need a deployment target like the current one: iOS 10.

However, to enable the `SwiftUI.Image` initializer, we'll need the iOS 13 deployment target. Yet, with such a deployment target, it won't be possible to import SFSafeSymbols for projects with a deployment target lower than iOS 13. There's a [post on the Swift forums](https://forums.swift.org/t/weak-linking-of-frameworks-with-greater-deployment-targets/26017) about this issue.

We should probably temporarily remove the `SwiftUI.Image` initializer and document why. But let's first await clarifications from the Swift Forums...

PR #26, opened 2019-06-21
Fix availability for Image extension

ericlewis:
This also include OSX, tvOS, watchOS in available targets for the SwiftUI Image extension. Closer matches how Apple themselves gate SwiftUI.

### DIFF
--- a/Sources/SFSafeSymbols/ImageExtension.swift
+++ b/Sources/SFSafeSymbols/ImageExtension.swift
@@ -2,12 +2,11 @@
 
 import SwiftUI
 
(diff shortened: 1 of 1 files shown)


fredpi:
Thanks, @ericlewis! 👍 We will probably still need to adjust the `Package.swift` to add proper platform support. There are also some other questions regarding that (see #24), but I’ll take care of them soon. Meanwhile, this can already be merged.

fredpi:
@ericlewis I've done the remaining work, should work properly now 👍

PR #30, opened 2019-06-25
Remove support for macOS

unnamedd:
As soon the initializer `Image(systemName:)` doesn't have support to macOS yet, in my point of view, doesn't make sense to support the platform yet.

Checking SwiftUI headers, specifically for the init method mentioned above, macOS/OSX is currently unavailable, for this reason I'm creating this pull request.

Here the snippet of code extracted from SwiftUI headers:
```swift
@available(iOS 13.0, tvOS 13.0, watchOS 6.0, *)
@available(OSX, unavailable)
extension Image {

    @available(OSX, unavailable)
    public init(systemName: String)
}
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -4,7 +4,7 @@ import PackageDescription
 
 let package = Package(
     name: "SFSafeSymbols",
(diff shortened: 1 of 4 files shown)


fredpi:
Thanks @unnamedd for noticing and fixing this! 👍

PR #34, opened 2019-09-18
SwiftUI is not available on older iOS versions.

jstarfruits:


fredpi:
@jstarfruits Thanks for the PR and sorry for not replying earlier – I'm short on time but will definitely have a look on Sunday!

fredpi:
@jstarfruits Thanks again for your PR 👍 – I released version `1.0.1` with your fix

PR #35, opened 2019-10-30
Fix macOS compilation

chriszielinski:
At the moment, Carthage fails when it attempts to build for the macOS platform. Since macOS is explicitly a supported platform (build settings), I added the appropriate attributes where necessary.


fredpi:
@chriszielinski Thanks for the PR – I'm short on time but will have a look on Sunday!

fredpi:
@chriszielinski Thanks again for your PR 👍 – I released version `1.0.1` with your fix

PR #36, opened 2019-12-03
Add CURRENT_PROJECT_VERSION

yonaskolb:
Hi, thanks for the project!

We're using this via Carthage and are running into an issue on submission to the app store as it comes back with the error:
```
SFSafeSymbols.framework is invalid. The Info.plist file is missing the required key: CFBundleVersion
```
The xcodeproj in this repo is generated via SwiftPM. SwiftPM currently has a bug in that it doesn't add a `CURRENT_PROJECT_VERSION` to the project https://bugs.swift.org/browse/SR-4265

This PR adds that `CURRENT_PROJECT_VERSION` by adding an `.xcconfig` file and re-generating with
```
swift package generate-xcodeproj --xcconfig-overrides Package.xcconfig
```

I could have just added the `CURRENT_PROJECT_VERSION` directly to the project, but I figured you might want to regenerate it later. Hence all the diffs as SwiftPM's generated project UUID's are not deterministic.


fredpi:
Thanks for the PR! 👍

fredpi:
@yonaskolb I have released [1.0.2](https://github.com/piknotech/SFSafeSymbols/releases/tag/1.0.2) with your fix

PR #39, opened 2019-12-30
Add support for UIApplicationShortcutIcon

sebj:
Hi, thanks for SFSafeSymbols!

I've noticed that it's possible to use SF Symbols for dynamic home screen quick action icons (passing them as Strings to `UIApplicationShortcutIcon`, similar to `UIImage` etc), so thought it'd be useful for SFSafeSymbols to have an initialiser extension for `UIApplicationShortcutIcon`.

Cloned the project with Xcode 11.3 and added the initialiser & a unit test.

For some reason, Xcode silently reformatted `project.pbxproj` with different formatting when I opened the project. This doesn’t seem to have caused any issues, but I'm open to any suggestions if this is undesired.

— Seb

### DIFF
--- a/Sources/SFSafeSymbols/Initializers/UIApplicationShortcutIconExtension.swift
+++ b/Sources/SFSafeSymbols/Initializers/UIApplicationShortcutIconExtension.swift
@@ -1,0 +1,16 @@
+#if canImport(UIKit) && (os(iOS) || targetEnvironment(macCatalyst))
+
+import UIKit
+
+@available(iOS 13.0, *)
+public extension UIApplicationShortcutIcon {
+
+    /// Create an icon using a system symbol image of the given type.
+    ///
+    /// - Parameter systemSymbol: The `SFSymbol` describing this image.
+    convenience init(systemSymbol: SFSymbol) {
+        self.init(systemImageName: systemSymbol.rawValue)
+    }
+}
+
+#endif
--- a/Tests/SFSafeSymbolsTests/UIApplicationShortcutIconExtensionTests.swift
+++ b/Tests/SFSafeSymbolsTests/UIApplicationShortcutIconExtensionTests.swift
@@ -1,0 +1,21 @@
+@testable import SFSafeSymbols
+
+#if os(iOS) || targetEnvironment(macCatalyst)
+
+import XCTest
+
+final class UIApplicationShortcutIconExtensionTests: XCTestCase {
+    func testSimpleInit() {
+        if #available(iOS 13.0, *) {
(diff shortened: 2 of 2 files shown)


SFSafeSymbols.xcodeproj/project.pbxproj
```diff
@@ -1,680 +1,553 @@
 // !$*UTF8*$!
```
fredpi: For me, Xcode doesn't do this – can you try to reset the changes to the project file and commit it (and then comment whether the project file still changes for you when opening Xcode?)

Sources/SFSafeSymbols/Initializers/UIApplicationShortcutIconExtension.swift
```diff
@@ -0,0 +1,19 @@
+#if canImport(UIKit) && !os(watchOS)
+
+import UIKit
+
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
```
fredpi: `@available(iOS 13.0, *)` should also work I guess

fredpi (changes requested):
Thanks for the PR 🚀👍 I just proposed two little things, would you mind having a look there?

SFSafeSymbols.xcodeproj/project.pbxproj
sebj: I'm pretty puzzled by Xcode's changes.. If I reset the project file and open the project, no changes are made, but the 2 new files are obviously missing. As soon as they're added to the project, the project file is reformatted in the same way as before, giving a huge diff. Using 11.3 (11C29)

fredpi: I just tried adding a new file to the project and also saw that reformatting – maybe the current project file style is "wrong"

I think it's fine then and I can merge this PR. Thanks again! 👍

sebj:
@fredpi Sorry – project should probably either be regenerated using SPM, or otherwise those two files added back through Xcode, as I reverted all of my `project.pbxproj` changes in the last commit 😬

fredpi:
@sebj Oh – no worries, I've fixed that in [this](https://github.com/piknotech/SFSafeSymbols/commit/fb1e7c23abe128be3f606f0700ad8ed6f99017a3) commit. Also, I have released a [new version](https://github.com/piknotech/SFSafeSymbols/releases/tag/1.1.1) with your changes.

sebj:
@fredpi Phew, thanks! Brilliant :)

Issue #14, opened 2019-06-05
Add UIButton initializer


fredpi:
System Images can also be used on UIButton, so we should provide an initializer for that, too

fredpi:
Closed via #40.

PR #40, opened 2020-03-23
Add UIButton initializer

posix88:
Hello and thanks for this project.
I've found the open issue #14  about adding some missing convenience methods for `UIButton`.

I've added the following two methods :
`func systemButton(with symbol: SFSymbol, target: Any?, action: Selector?) -> Self`
`func setImage(with symbol: SFSymbol, for state: UIControl.State)`

that leverage on the `UIImage` extension and `UIButton` native methods.

I've also inserted some tests for the new behaviors.

Greetings


fredpi:
Thank you very much, @posix88! 🚀

PR #51, opened 2020-08-18
fix: Make eyedropper.halffull use correct symbol

Robuske:
`eyedropper.halffull` was using `eyedropper.full` symbol for its documentation

(GitHub does not render it, but if you copy from the SFSymbols app and use safari search it detects the symbol)


fredpi:
Thanks @Robuske 🎉

PR #58, opened 2020-10-27
Make SwiftUI Label extension public

sebj:
Fixes a mistake in the SwiftUI `Label` extension I previously contributed – it wasn't `public` 🤦‍♂️


fredpi:
Thanks @sebj!

PR #60, opened 2020-11-22
SFSymbols 2.1

fredpi:
This is currently not ready to get merged, as an OS-version-specific handling must be done for symbols that aren't there despite the symbols app saying they are available.


fredpi:
Turns out all symbols said to be available are actually available

PR #61, opened 2020-12-08
Add dynamic target

noppefoxwolf:


fredpi:
Thanks for the PR, @noppefoxwolf! 👍

Could you explain for which use case the dynamic library is needed? I don't have much knowledge about static / dynamic libraries and would like to gain some understanding about why this is needed before merging it.

noppefoxwolf:
Thank you.
I'm not good at English.
So, I pushed example repository.
https://github.com/noppefoxwolf/DynamicUsecase

This application dependent two frameworks `FrameworkA` and `FrameworkB`.
Both frameworks dependent same static framework using SwiftPM.

As you can see from the build error, Xcode cannot link duplicated libraries.
```
Swift package product 'SFSafeSymbols' is linked as a static library by 'DynamicUsecase' and 2 other targets. This will result in duplication of library code.
```

To solve this, library needs to link as dynamic.

fredpi:
@noppefoxwolf Thanks for providing an example! Makes sense to me to offer a dynamic variant of SFSafeSymbols. I will now merge your PR and publish a new release 👍

PR #62, opened 2020-12-17
Mark UIImage extension initializers as available for watchOS

tomasf:
The UIImage initializers in UIImageExtension.swift were marked as unavailable for watchOS, but the underlying `UIImage(systemName:[...])` initializers do exist and work on watchOS.


fredpi:
Looks good to me, thanks @tomasf 👍

PR #64, opened 2021-02-18
fix Label.init(_:systemImage:) type inference issue

### LINKED ISSUE #63: StringProtocol type inference issue in SwiftUI's Label initializers
I just noticed a strange behavior. In `SwiftUILabelExtension`, the protocol `StringProtocol` is always matched to `String` instead of `LocalizedStringKey` if you write a string literal in your code. This is inconsistent with other SwiftUI native methods. For example:

```swift
// Native SwiftUI:
Text("some text")                      // inferred as LocalizedStringKey
Label("some text", systemImage: "...") // inferred as LocalizedStringKey
    .navigationBarTitle("some text")   // inferred as LocalizedStringKey

// SwiftUILabelExtension:
Label("some text", systemSymbol: .xxx) // hmm... inferred as String... 🤔
```

So I have to do the following to ensure the text is being localized correctly:

```swift
Label("some text" as LocalizedStringKey, systemSymbol: .xxx) // 🤔
```

I checked the implementation of `SwiftUILabelExtension` and it seems to be defined in the same way as in SwiftUI's native methods, but I don't know why the compiler doesn't give priority to matching `LocalizedStringKey`.

```swift
public extension Label where Title == Text, Icon == Image {
    
    /// Creates a label with a system symbol image and a title generated from a
    /// localized string.
    ///
    /// - Parameter systemSymbol: The `SFSymbol` describing this image.
    init(_ titleKey: LocalizedStringKey, systemSymbol: SFSymbol) {
        self.init(titleKey, systemImage: systemSymbol.rawValue)
    }
    
    /// Creates a label with a system symbol image and a title generated from a
    /// string.
    ///
    /// - Parameter systemSymbol: The `SFSymbol` describing this image.
    init<S>(_ title: S, systemSymbol: SFSymbol) where S : StringProtocol {
        // 🤔 always goes here with S==String...
        self.init(title, systemImage: systemSymbol.rawValue)
    }
}
```

I suggest to check if we can find a better way to define the initializer signatures. Because this inconsistent behavior can easily lead to localization bugs, and it is less easy to be found if the app is not fully tested.

ddddxxx:
fix #63

### DIFF
--- a/Sources/SFSafeSymbols/Initializers/SwiftUI/SwiftUILabelExtension.swift
+++ b/Sources/SFSafeSymbols/Initializers/SwiftUI/SwiftUILabelExtension.swift
@@ -9,7 +9,6 @@ public extension Label where Title == Text, Icon == Image {
     /// localized string.
     ///
     /// - Parameter systemSymbol: The `SFSymbol` describing this image.
-    @_disfavoredOverload
     init(_ titleKey: LocalizedStringKey, systemSymbol: SFSymbol) {
         self.init(titleKey, systemImage: systemSymbol.rawValue)
     }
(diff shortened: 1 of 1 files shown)


fredpi:
Wow, great catch, @ddddxxx! 🎉

PR #67, opened 2021-02-24
Add back CURRENT_PROJECT_VERSION

yonaskolb:
#36 fixed app store deployment but #54 unfortunately broke it again.
This PR simply adds back the `CURRENT_PROJECT_VERSION` but doesn't add the `Package.xcconfig` as I'm not sure how you want to maintain your xcode project moving forward. Is it still generated via SPM or maintained by hand, as generation seems to be broken due to https://github.com/piknotech/SFSafeSymbols/issues/68


fredpi:
@yonaskolb Thank you very much for the PR!

I'm sorry for breaking app store deployment again and I don't know why I did so – in https://github.com/piknotech/SFSafeSymbols/commit/b458283c6e51e903fcd87f76cd865a22c32ec702 I restructured the Xcode project and decided not to auto-generate the `.xcodeproj` in the future, rather maintaining it manually. I don't understand where exactly `CURRENT_PROJECT_VERSION` was removed – I can't even find where it was in added, as #36 seemingly only adjusted the `Package.xcconfig`, but didn't add `CURRENT_PROJECT_VERSION = 1;` to the `SFSafeSymbols.xcodeproj/project.pbxproj` file, as this PR does.

But now, that this PR readds it and the `.xcodeproj` is maintained manually, I'm confident it won't happen again!

PR #71, opened 2021-03-15
fix string initializer

### LINKED ISSUE #66: Many cases can't be initialized from raw value
In many cases, for example `15.square.fill`, the rawValue returned from its enum case (`SFSymbol._15SquareFill`) will return the proper name (`15.square.fill`), but it cannot be initialized from its rawValue (`SFSymbol(rawValue: 15.square.fill)` returns nil).

ddddxxx:
Currently `SFSymbol` can only be initialized from rather unexpected string:
```swift
SFSymbol(rawValue: "_0Circle") // OK
SFSymbol(rawValue: "0.circle") // nil
```

This pr change it to more intuitional behavior:
```swift
SFSymbol(rawValue: "_0Circle") // nil
SFSymbol(rawValue: "0.circle") // OK
```

This is breaking change and probably need major version bump.

fix #66


ddddxxx:
Giving every symbol a fixed `rawValue` breaks renamed symbols. Previously a renamed symbol have the same availability as the old symbol, and returns old symbol name on old system, which requires massive `switch` in `rawValue`.

Bump system requirement for these symbols will make them almost unusable:
```diff
- @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
+ @available(iOS 14.2, macOS 11.0, tvOS 14.2, watchOS 7.1, *)
  case character = "character"
```

I don't think it can be fixed in current design. Once we make `SFSymbol` enum, we can do something like this:
```swift
@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
static let character: SFSymbol = {
    if #available(iOS 14.2, macOS 11.0, tvOS 14.2, watchOS 7.1, *) {
        return SFSymbol(rawValue: "character")
    } else {
        return SFSymbol(rawValue: "a")
    }
}()
```

What do you think @fredpi ?

fredpi:
@ddddxxx Thanks for the PR! While we can't merge it because of the issues you discovered, it helps with the understanding of the problem.

I commented in #66 because it's about the same issue (`rawValue` initialization). As we figured there, a change is most probably only possible by migrating to a `struct` instead of an `enum`, just as you suggested in your last comment.

I don't remember exactly, but I think the reason for the large `switch` to retrieve the `rawValue` (instead of specifying it directly) was that I encountered the same issue with name versions as you did in this PR.

PR #86, opened 2021-11-27
Improve package specs

### LINKED ISSUE #78: Issues with static_framework = true in SFSafeSymbols.podspec
Is there a reason, why the _SFSafeSymbols.podspec_ contains the `static_framework = true` line?

I try to use SFSymbols in a larger project, where SFSafeSymbols is needed in a private framework (and private Cocoapod). As soon as I add SFSafeSymbols as dependency to the private frameworks podspec file, I'll get an error when I try to `pod update` the main project:

`… target has transitive dependencies that include statically linked binaries: (SFSafeSymbols)`

I tried to comment out the `static_framework = true` line in a fork of SFSafeSymbols and this solves the problem. But I'd prefer to set the dependency to the main repo of SFSafeSymbols and not needing to maintain the fork…

So is there an important reason for the `static_framework = true` in the first place?
Or is there another solution for the issue (and no, I cannot just switch to SPM for many reasons)?

StevenSorial:
This PR improves the package specs (`Package.swift` and `SFSafeSymbols.podspec`):

* General: Lower the library requirements to 2016 platforms (iOS 10.0, macOS 10.12, tvOS 10.0, watchOS 3.0).
* `Package.swift`: Raise the minimum Swift version to `5.3` to match `SFSafeSymbols.podspec`.
* `Package.swift`: Remove the library type (static/dynamic) which lets SPM picks it automatically ([recommended by Apple](https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#product)). Tested successfully on the repo provided in #61.
* `SFSafeSymbols.podspec`: Add support for swift `5.4` and `5.5`.
* `SFSafeSymbols.podspec`: Remove `static_framework = true`. Fixes #78.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,13 +1,12 @@
-// swift-tools-version:5.1
+// swift-tools-version:5.3
 
 import PackageDescription
 
 let package = Package(
     name: "SFSafeSymbols",
-    platforms: [.iOS(.v11), .tvOS(.v11), .watchOS(.v6), .macOS(.v10_13)],
+    platforms: [.iOS(.v10), .tvOS(.v10), .watchOS(.v3), .macOS(.v10_12)],
     products: [
-        .library(name: "SFSafeSymbols", type: .static, targets: ["SFSafeSymbols"]),
-        .library(name: "SFSafeSymbols-Dynamic", type: .dynamic, targets: ["SFSafeSymbols"])
+        .library(name: "SFSafeSymbols", targets: ["SFSafeSymbols"]),
     ],
     dependencies: [],
     targets: [
--- a/Package.swift
+++ b/Package.swift
@@ -4,7 +4,7 @@ import PackageDescription
 
 let package = Package(
     name: "SFSafeSymbols",
-    platforms: [.iOS(.v10), .tvOS(.v10), .watchOS(.v3), .macOS(.v10_12)],
+    platforms: [.iOS(.v11), .tvOS(.v11), .watchOS(.v4), .macOS(.v10_13)],
     products: [
         .library(name: "SFSafeSymbols", targets: ["SFSafeSymbols"]),
     ],
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ The following SF Symbols versions are currently supported:
 
 `SFSafeSymbols` can be installed via the **Swift Package Manager (recommended)**, Carthage or CocoaPods.
 
-Supported platforms are `iOS (10.0+)`, `macOS (10.12+)`, `tvOS (10.0+)` and `watchOS (3.0+)`, although the actual functionality is of course only accessible starting with `iOS 13.0`, `macOS 11.0`, `tvOS 13.0` and `watchOS 6.0`.
+Supported platforms are `iOS (11.0+)`, `macOS (10.13+)`, `tvOS (11.0+)` and `watchOS (4.0+)`, although the actual functionality is of course only accessible starting with `iOS 13.0`, `macOS 11.0`, `tvOS 13.0` and `watchOS 6.0`.
 
 ### Swift Package Manager (Xcode-integrated)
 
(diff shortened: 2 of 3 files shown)


fredpi:
@StevenMagdy Thank you!! I like the changes but have some questions:

> * General: Lower the library requirements to 2016 platforms (iOS 10.0, macOS 10.12, tvOS 10.0, watchOS 3.0).

Have you tested whether everything works fine when the repo is used for an iOS 10 app which is then run on an iOS 10 simulator? I don't remember the reason why I picked iOS 11.0 as the minimum deployment target in the first place and want to make sure this was in fact an arbitrary choice.

> * `Package.swift`: Remove the library type (static/dynamic) which lets SPM picks it automatically ([recommended by Apple](https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#product)). Tested successfully on the repo provided in [Add dynamic target #61](https://github.com/SFSafeSymbols/SFSafeSymbols/pull/61).

Great catch!

> * `SFSafeSymbols.podspec`: Add support for swift `5.4` and `5.5`.

What does this added support mean? Does it enable something that has not been possible previously or is this rather a decorative change?

> * `SFSafeSymbols.podspec`: Remove `static_framework = true`. Fixes [Issues with static_framework = true in SFSafeSymbols.podspec #78](https://github.com/SFSafeSymbols/SFSafeSymbols/issues/78).

Is this change similar to the change in the `Package.swift`, i. e. letting the system automatically pick whether the framework is static or dynamic? Or does dropping `static_framework = true` mean that the framework is always dynamic? I want to make sure in order not to annoy our CocoaPods users.

StevenSorial:
> Have you tested whether everything works fine when the repo is used for an iOS 10 app which is then run on an iOS 10 simulator?

yes, I did a small test. I don't see why it shouldn't work since all of the APIs have a minimum of iOS 13.0.

>  I don't remember the reason why I picked iOS 11.0 as the minimum deployment target in the first place and want to make sure this was in fact an arbitrary choice.

what was weirder for me, is that all the platforms had the minimum of 2017 platforms, except watchOS was 6.0 (2019). Libraries typically set the minimum to the platforms of the same year.

> What does this added support mean? Does it enable something that has not been possible previously or is this rather a decorative change?

yes, I think it became decorative after Swift gained ABI Stability in 5.0

> Is this change similar to the change in the Package.swift, i. e. letting the system automatically pick whether the framework is static or dynamic? Or does dropping static_framework = true mean that the framework is always dynamic? I want to make sure in order not to annoy our CocoaPods users.

I think it's different from `Package.swift`, because the `Podfile` still can specify one type over the other using `use_frameworks!`.

This is not a topic that I'm very familiar with. I just took a look at the specs of some famous Swift libraries that are probably used in the most complex projects (Alamofire, RxSwift, SnapKit, and Lottie). None of them have `static_framework = true`.

fredpi:
@StevenMagdy I tracked down why I decided to use iOS 11 as the minimum deployment target – it's because CocoaPods failed with iOS 10 as the deployment target: https://github.com/CocoaPods/CocoaPods/issues/8915#issuecomment-506326496

The error (which prevents CocoaPods pushes of new releases) can be seen when running `pod lib lint --no-clean`.

Therefore I suggest to change back to iOS 11 (but keep the update to the matching watchOS version).

Normally, the CI would indicate the CocoaPods lint failure, but it seems like the CI has not been triggered by this PR. I will try to fix that now ;)

fredpi:
Turns out the CI has run, but the run has not been linked back to this PR: https://app.bitrise.io/build/a274e826-a7b1-4287-a3ab-471bee0720f2#?tab=log

StevenSorial:
@fredpi I think it's trying to hook to the old organization

fredpi:
@StevenMagdy Maybe it's fixed now, but I'm not sure about it. Maybe we can test by creating a new dummy PR...

This PR can be merged, if the CI passes: https://app.bitrise.io/build/e3619c6a-d9ea-45ce-9ddb-98e478c68be1#?tab=log

fredpi (approved):

fredpi:
CI has passed; merging 🎉

PR #91, opened 2021-12-11
Fix macOS availability

StevenSorial:
macOS availability for version 1.0 in `name_availability.plist` is 10.15. but initializers for both [Image](https://developer.apple.com/documentation/swiftui/image/init(systemname:)) and [NSImage](https://developer.apple.com/documentation/appkit/nsimage/3622472-init) actually has requirement of 11.0, also `SFSymbols.app` has macOS minimum for version 1.0 as 11.0.

The reason for why it's written as 10.15 in `name_availability.plist` is unknown but if I have to guess, the font was privately available in 10.15 for catalyst apps compatibility. Whatever the reason is, we should have the same requirements as the initializers.


StevenSorial:
@fredpi @knothed  Should I move forward merging this?

fredpi:
@StevenMagdy Sorry for the late reply!

Thanks for the PR; looks good to me 👍

Issue #93, opened 2022-01-06
Unit tests for correct symbols


MartinW:
Add unit tests to check that the correct symbol is being retrieved using XCTAssert. Compare Image(systemName:) value against Image(systemSymbol:) value. Currently it only checks that it has not errored.

- Add sorting to test symbols in alphabetical order
- Add version tests
- Add success and fail tests for specific symbols

fredpi:
Closed via #94

PR #94, opened 2022-01-06
Unit test refactoring

MartinW:
Suggested fixes for #93


fredpi:
@MartinW Thank you very much for this PR and sorry for the delay! I agree with you, that semantically, it is better to test the way that you suggested. I will now merge this and then create a follow-up PR which also adjusts the other tests similarly.

fredpi (approved):

PR #98, opened 2022-03-23
Support explicit symbol localization

knothed:
Implements #87.

Follows [this](https://github.com/SFSafeSymbols/SFSafeSymbols/pull/90#issuecomment-1010614253) suggestion by @StevenMagdy for the static localization and [this](https://github.com/SFSafeSymbols/SFSafeSymbols/pull/90#issuecomment-1013672769) by @StevenMagdy for the dynamic localization.


knothed:
Tomorrow I'll implement some minor remaining stuff, including documenting the localization features in the REAMDE.

I'll request your review then @SFSafeSymbols/core-contributors

knothed:
Done. Please review.

I have incorporated all of the things which have been remarked in the review of #90 in this MR.

README.md
```diff
@@ -167,6 +167,38 @@ NSImage(systemSymbol: .cCircle)
...
+
+  Static localization only exposes the localizations which are actually available, so you cannot accidentally localize a non-localizable symbol.
+
+  Dynamic localization is useful however when dealing with an array of `LocalizableSFSymbols` which all have different localization information.
+
+Attention: Serializing and deserializing `SFSymbol`s makes them lose their explicit localization information. This is because `SFSymbol(rawValue:)` creates a standard, non-localizable `SFSymbol`, whereas `SFSymbol.character` and all other static symbols have type `LocalizableSFSymbol: SFSymbol` which doesn't support custom serialization yet.
+
+
+
+To use them ine the code, there are multiple options:
```
fredpi: Small typo here

README.md
```diff
@@ -167,6 +167,38 @@ NSImage(systemSymbol: .cCircle)
...
+  // 2. Dynamic localization:
+  SFSymbol.character.availableLocalizations // [.ar, .he, .hi, .ja, .ko, .th, .zh, .zhTraditional]
+  let a = SFSymbol.character.localized(to: .ar)
+  let b = SFSymbol.character.localized(to: .rtl)
+  // a, b have type SFSymbol?
+  ```
+
+  Static localization only exposes the localizations which are actually available, so you cannot accidentally localize a non-localizable symbol.
+
+  Dynamic localization is useful however when dealing with an array of `LocalizableSFSymbols` which all have different localization information.
```
fredpi: I would rather write `Dynamic localization, in contrast, is useful...`.

Also `localization information` sounds a bit weird. Maybe better use `available localizations`.

README.md
```diff
@@ -167,6 +167,38 @@ NSImage(systemSymbol: .cCircle)
 NSImage(systemSymbol: SFSymbol.eCircleFill, accessibilityDescription: "some.description")
 ```
 
+## Localization
+
+SFSymbols can come with multiple different localizations. `SFSafeSymbols` exposes localization the following way:
+
+- Implicit localization: When using an `SFSymbol`, it gets automatically localized to the user's current locale - nothing to do on your part.
```
fredpi: I would suggest to explain that this behavior is not something that is done by `SFSafeSymbols`, but rather by the system itself already. For example using the sentence `This behavior is managed by Apple's system implementation of SF Symbols.`

README.md
```diff
@@ -167,6 +167,38 @@ NSImage(systemSymbol: .cCircle)
 NSImage(systemSymbol: SFSymbol.eCircleFill, accessibilityDescription: "some.description")
 ```
 
+## Localization
+
+SFSymbols can come with multiple different localizations. `SFSafeSymbols` exposes localization the following way:
```
fredpi: `SFSymbols` → `SF Symbols` (that's the official name)

Sources/SFSafeSymbols/Symbols/LocalizableSFSymbol.swift
```diff
@@ -0,0 +1,203 @@
...
+        } else {
+            return nil
+        }
+    }
+}
+
+// MARK: Static Localization
+
+@available(iOS 13.0, macOS 11.0, tvOS 13.0, watchOS 6.0, *)
+public class NonLocalizedSymbol: LocalizableSFSymbol {
```
fredpi: `NonLocalizedSymbol: LocalizableSFSymbol` 🤣

I don't have any better ideas regarding naming, but wanted to point out how funny this is 😅 Let's hope Variadic Generics come soon...

Sources/SFSafeSymbols/Symbols/SFSymbol+AllSymbols.swift
```diff
@@ -14,9 +14,9 @@ extension SFSymbol {
             return allSymbols3_0
         } else if #available(iOS 14.5, macOS 11.3, tvOS 14.5, watchOS 7.4, *) {
             return allSymbols2_2
-        } else if #available(iOS 14.2, macOS 11.0, tvOS 14.2, watchOS 7.1, *) {
+        } else if #available(iOS 14.2, tvOS 14.2, watchOS 7.1, *) {
```
fredpi, line 17: Is this correct? Shouldn't the 2.1 symbols be available for macOS 11.0?

SymbolsGenerator/Sources/SymbolsGenerator/Helpers/×.swift
```diff
@@ -0,0 +1,10 @@
+infix operator ×
```
fredpi, line 1: 😜

fredpi (changes requested):
@knothed Thank you very much! 🎉

I just had a look over your changes and am very happy with the result and the code quality! I left some minor remarks, mostly addressing things in the README. Maybe you can have a quick look over them so that this PR can be quickly merged.

fredpi:
@StevenMagdy Do you have any objections about merging this soon? Merging this will finally allow us to finish up the 3.2 release (we should probably release it as 3.2 to stay in-line with Apple's versioning).

StevenSorial:
@knothed @fredpi Sorry for taking so long. I'm going to take a look next weekend.

Sources/SFSafeSymbols/Symbols/SFSymbol+AllSymbols.swift
knothed: It is: I removed availability annotations which are not more restrictive than the base availability (i.e. macOS 11.0). The outer (i.e. function's) availability already has macOS 11.0 specified. This removes warnings which came when building on macOS.

README.md
fredpi: Did you read the second comment? (_"Also `localization information` sounds a bit weird. Maybe better use `available localizations`."_) Because if you don't have anything against it, I would suggest to change the wording.

fredpi:
@knothed This PR now has some conflicts with the new tests as merged in #94. I will soon push a new PR with even more changes to the tests, and after that PR is merged, we can rebase this.

fredpi:
@knothed I rebased your branch to the current `stable`.

@StevenMagdy No need to be sorry! After all we have taken way too much time ourselves... And if you don't have the time to take an in-depth look: After my analysis, I'm confident that this implementation works well 👍

fredpi (approved):

Sources/SFSafeSymbols/Symbols/LocalizableSFSymbol.swift
```diff
@@ -0,0 +1,203 @@
...
+public protocol SymbolLocalization {
+    init(source: SFSymbol)
+}
+
+// TODO: replace the following with Variadic Generics once available
+// Since the code is currently only an interim solution, we don't auto-generate it.
+// Beware: if more localizations are added to SFSymbols before variadic generics are introduced to Swift, you may have to create the missing specialized classes.
+
+@available(iOS 13.0, macOS 11.0, tvOS 13.0, watchOS 6.0, *)
+public /* abstract */ class LocalizableSFSymbol: SFSymbol {
```
fredpi: What do you think: Maybe we can name this `SFSymbolWithLocalizationInformation` instead of `LocalizableSFSymbol`...

`LocalizableSFSymbol` is somehow misleading. I thought about this when reading your comment in #100: What will the user think when they are presented a `Set<LocalizableSymbol`...

SymbolsGenerator/Sources/SymbolsGenerator/Helpers/Types.swift
```diff
@@ -39,12 +39,32 @@ struct Availability: Comparable, Equatable, Hashable {
...
+    static private(set) var base: Availability!
+
+    /// Convert into an expression than can be used in code when prefixed with either `#` or `@`.
+    var availableExpression: String {
+        "available(iOS \(iOS), macOS \(macOS), tvOS \(tvOS), watchOS \(watchOS), *)"
+    }
+
+    /// Convert into an expression than can be used in code when prefixed with either `#` or `@`.
+    /// Availability specifications which are not strictly more restrictive than `Availability.base` are dropped.
+    var availableExpressionWithoutRedundancyToBase: String {
```
StevenSorial, line 51: Nice!

Sources/SFSafeSymbols/Symbols/SymbolLocalizations.swift
```diff
@@ -0,0 +1,330 @@
...
+    case zh = "zh"
+    case zhTraditional = "zh.traditional"
+}
+
+@available(iOS 13.0, macOS 11.0, tvOS 13.0, watchOS 6.0, *)
+internal extension LocalizableSFSymbol {
+    var _availableLocalizations: Set<Localization> {
+        var result = Set<Localization>()
+        if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *) {
+            if (localizations.contains { $0 == Ar_v3.self }) { result.update(with: .ar) }
```
StevenSorial, line 23: `localizations` is allocated and released too many times inside this scope, because it's a computed property. Can we cache it at the top?

Sources/SFSafeSymbols/Symbols/SymbolLocalizations.swift
```diff
@@ -0,0 +1,330 @@
...
+            if (localizations.contains { $0 == Ar_v2.self }) { result.update(with: .ar) }
+            if (localizations.contains { $0 == He_v2.self }) { result.update(with: .he) }
+            if (localizations.contains { $0 == Hi_v2.self }) { result.update(with: .hi) }
+            if (localizations.contains { $0 == Ja_v2.self }) { result.update(with: .ja) }
+            if (localizations.contains { $0 == Ko_v2.self }) { result.update(with: .ko) }
+            if (localizations.contains { $0 == Rtl_v2.self }) { result.update(with: .rtl) }
+            if (localizations.contains { $0 == Th_v2.self }) { result.update(with: .th) }
+            if (localizations.contains { $0 == Zh_v2.self }) { result.update(with: .zh) }
+        }
+        if #available(*) {
```
StevenSorial, line 52: Didn't know that was possible, nice!
What does it achieve? Is it just for consistency with the other `if`s?

Sources/SFSafeSymbols/Symbols/SymbolLocalizations.swift
```diff
@@ -0,0 +1,330 @@
...
+    public var zh: SFSymbol { .init(rawValue: "\(source.rawValue).\(Localization.zh.rawValue)") }
+}
+
+@available(iOS 13.0, macOS 11.0, tvOS 13.0, watchOS 6.0, *)
+public struct Zh_v3: SymbolLocalization {
+    let source: SFSymbol
+    public init(source: SFSymbol) { self.source = source }
+    @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+    public var zh: SFSymbol { .init(rawValue: "\(source.rawValue).\(Localization.zh.rawValue)") }
+}
```
StevenSorial: Nit: can you add a new line at the end? Just opening the file makes Git register a change because some editors add a new line automatically.

Sources/SFSafeSymbols/Symbols/SFSymbol+AllSymbols.swift
```diff
@@ -5,7 +5,7 @@ extension SFSymbol {
     @available(*, deprecated, renamed: "allSymbols")
     public static var allCases: [SFSymbol] { Array(allSymbols) }
```
StevenSorial: Should we make this also `[LocalizableSFSymbol]`?

Sources/SFSafeSymbols/Symbols/LocalizableSFSymbol.swift
StevenSorial: @fredpi @knothed I don't think `LocalizableSFSymbol` and `NonLocalizedSymbol` serve us any benefit here. I think `SFSymbol` by itself will mean that it does not have *static* localization (= NonLocalizedSymbol), and `LocalizableSFSymbol` properties/functions can also be safely moved to `SFSymbol`.  Am I missing something?

StevenSorial (dismissed):
@knothed Thank you for the huge effort 🎉 
I added a few minor remarks for the current approach as-is and will discuss how we can improve the approach itself here.

### Avoid deriving the runtime localization from the static localization

As I mentioned in #90, I think the property `localizations: [SymbolLocalization.Type]` should only be used in `internal` code and tests, implementing the runtime lozalization in this way is only beneficial when type-erasing symbols provided by `SFSafeSymbols` but will fall short when initializing `LocalizableSFSymbol` manually using a `String` or `Codable`. I think its confusing when `LocalizableSFSymbol.character.availableLocalizations` and `LocalizableSFSymbol(rawValue: "character").availableLocalizations`, have different results.
i think the runtime localization should be implemented using:

```swift
extension SFSymbol {
    internal static let localizationDictionary: [SFSymbol: Set<Localization>] = { ... }()

    public func has(localization: Localization) -> Bool {
        return Self.localizationDictionary[self]?.contains(localization) ?? false
        // or UIImage(systemName: "\(rawValue).\(localization.rawValue)") != nil 😜
    }
}
```
`localizationDictionary` will be huge and will increase the library size by a considerable amount, but if we could be smart about it, we would get rid of the `allSymbols` code to be `allSymbols = Set(localizationDictionary.keys)` effectively bringing the size back to how it is now.

### Going runtime-only

I think we've discussed this before but it's worth repeating. I think providing both runtime and static APIs is redundant and confusing, especially if the above modification is implemented. I don't think that optionality in `localized(to:)` is too bad to the point that we have to make a whole different API for it. Symbol (explicit) localization is a niche feature and even most developers don't know about it, so we should keep its API(s) and implementation as simple as possible.

even though I'm the one who suggested the generic API. 😄

Sources/SFSafeSymbols/Symbols/SFSymbol+AllSymbols.swift
fredpi: I propose not to, because it's the deprecated variant of `allSymbols` and as such should probably keep its type from the previous version

Sources/SFSafeSymbols/Symbols/SymbolLocalizations.swift
fredpi: I think it's for consistency only 😅

Personally, I have no preferences on whether to use ` if #available(*) {` or just put the code without an if

Sources/SFSafeSymbols/Symbols/SymbolLocalizations.swift
fredpi: Sounds reasonable (although maybe the compiler optimizes it that way anyway)

fredpi:
@StevenMagdy Thanks for the review!

### Dictionary for runtime localization

I like the idea of the dictionary for runtime localization as it will solve the problem of losing localization information. A few remarks:

- I see the disadvantage that we get a mismatch between runtime localization and static localization (which I would like to keep; see below)
- I'm wondering how it will work out performance-wise? At least for the first time (we can make it lazy), a dictionary lookup is needed...
- Implementing runtime localization this way will probably also allow us to merge `LocalizableSymbol` and `SFSymbol`. The separation made some sense before to differentiate between symbols that have localization information available and those that didn't (for example because they were decoded). But maybe it's still sensible, because static localization is still not available for custom created symbols...
- It will also be quite easy to implement, as we can just build upon the existing AllSymbols generation infrastructure.

### Dropping static localization entirely

Regarding your suggestion to drop static localization entirely: I understand that that may be more clear to the user, but runtime localization has the disadvantage that one cannot safely get the localization:

E. g. if a user writes `character.ar`, and support for `.ar` is removed in an upcoming release (unlikely; but who knows...), the user will know because they will get a compile error. In contrast, if the user wrote `character.localized(to: .ar)` in the first place, having confirmed previously that `.ar` localization is available by having a look at the symbol documentation, they will only get a runtime problem (i. e. a `nil` symbol), but not a compile error – and that's somehow against the idea of this repository...

Sources/SFSafeSymbols/Symbols/SymbolLocalizations.swift
knothed: Even if you had a preference, it **really doesn’t** matter - after all, it’s generated code the user will never come to see anyways.

Yes, it’s a noop.

knothed:
>  I don't think that optionality in localized(to:) is too bad to the point that we have to make a whole different API for it. Symbol (explicit) localization is a niche feature and even most developers don't know about it, so we should keep its API(s) and implementation as simple as possible.

That‘s true. I don’t believe that there is any use case which justifies that the user must statically know whether a specific localization is available or not. If the user wants to localize something explicitly, dynamic localization plus fall-back to the non-localized symbol should suffice. While I really like the variadic-generics-implementation of static localization, it is probably confusing to the novice user. Whether that is a bad thing however is debatable.

What I **really like** about static localization is how it looks to the library user in Xcode. On writing `SFSymbol.character.`, the user can just use autocomplete to choose between the available localizations on a per-symbol basis. This feature perfectly aligns with the main goal of SFSafeSymbols which is to mimick the simple syntax of SF Symbols itself.
Moreover, writing `SFSymbol.character.ar` doesn’t just look better than `SFSymbol.character.localized(to: .ar) ?? SFSymbol.character`, but conveys a premium feeling, as if the library was written by Apple itself: no need for detours, just straight to the point.

> Symbol (explicit) localization is a niche feature

Yes, and we’ve made far too big a fuss about it 😜

StevenSorial:
I agree, releasing a new version is more important anyway 😄

Sources/SFSafeSymbols/Symbols/LocalizableSFSymbol.swift
knothed: You're right, we don't need the distinction between `SFSymbol` and `LocalizableSFSymbol`. I'll change it.

StevenSorial:
@knothed Thanks for the changes. Are we going to change the runtime API to be Dictionary-based?

knothed:
@StevenMagdy I‘d propose to leave it as is for now and finally make a release. We can still think about and implement the dictionary approach afterwards in a non-sourcebreaking way. But there is no rush to it, it’s really a fancy niche feature.

StevenSorial:
> But there is no rush to it

I'm ok with that. we can always improve it in a later PR or a minor release.

StevenSorial (approved):

fredpi:
Thanks everyone! Merging...

Issue #97, opened 2022-02-19
Deprecated symbol names?


LilaQ:
Hi, I'm using SFSafeSymbols in my new project, but I get +100 Warnings in my log that several names were deprecated / changed in iOS 14. 

Is there a way to get a release that's for iOS 14+ ? 
I installed through CocoaPods btw.

Thanks in advance

knothed:
Hi @LilaQ, thanks for raising this issue!

Just to clarify: Are you seeing compile-time warnings or runtime warnings (i. e. in the log)? Could you maybe provide a screenshot that shows the warnings that you are seeing?

abdelmagied94:
Hi @knothed,

I'm also facing this issue. I'm using the package in a fresh project with a deployment target `iOS 14.0`.

<img width="1181" alt="Screen Shot 2022-04-06 at 11 58 25 PM" src="https://user-images.githubusercontent.com/48062713/162078948-4b03ba20-37e2-44a2-a4f2-a328fef7d5ab.png">

fredpi:
This problem is probably already fixed in the current `stable` and a release will be out soon.

Maybe you can give the current `stable` a try to test whether it really works, @abdelmagied94?

abdelmagied94:
@fredpi Thank you for your response 🙏  I found the issue that I delete the `IPHONEOS_DEPLOYMENT_TARGET` from pods projects as follows:
```
post_install do |installer|
  installer.pods_project.targets.each do |target|
    target.build_configurations.each do |config|
      config.build_settings.delete 'IPHONEOS_DEPLOYMENT_TARGET'
    end
  end
end
```

StevenSorial:
@abdelmagied94 if your deployment target is 14.0, then you can get rid of the warnings by applying the fix-it and upgrading to the new names. If I understood correctly this behavior is intended ([read](https://github.com/SFSafeSymbols/SFSafeSymbols/pull/80#issuecomment-974787421)) because its showing upgrades that are now possible because you have 14.0 as the minimum requirement.

@fredpi please correct me if I got it wrong

fredpi:
@StevenMagdy I also thought that at first, but in fact these are warnings from the library itself – to be precise: from the `allCases` array. In previous versions, we would include deprecated symbols in that array and that is now fixed.

StevenSorial:
I should've looked at the filename before answering 😄 

[...]

PR #101, opened 2022-04-09
Reduce redundancy while constructing `allSymbols`

StevenSorial:
This removes redundant symbols for `allSymbols` to reduce code/package size.


fredpi:
@StevenMagdy Thanks for the PR!

Two quick thoughts:

- While the "calculation" of the `allSymbols` set takes place only once per runtime, I'd still be interested in how long it takes compared to the previous variant. Maybe you can measure it?
- This implementation currently doesn't take care of removed symbols. This is e. g. the case with deprecated symbols. Without removing them, we get duplicate symbols and issues like #97.

StevenSorial:
> This implementation currently doesn't take care of removed symbols.

You are totally right. I completely forgot the whole reason of the current approach. 🤦🏼‍♂️ 

Closing.

knothed:
There is no need to close this. Just generate an extra array of removed symbols for every version which are then removed from the created allSymbols set.

StevenSorial:
@knothed Good idea. I wonder if that is going to be less code than declaring the whole Set as we currently do.

knothed:
I believe so as there are not _that_ many deprecated symbols compared with how many there are in total.

StevenSorial:
Removed deprecated symbols by declaring another set for each availability as suggested by @knothed 

@fredpi I did some quick measuring, and it did increase from 2ms to 3ms in an unoptimized build. I think that is ok considering that we decreased the size of the code, and it's only calculated once anyway.

knothed (approved):

knothed:
Great, looks good!

StevenSorial:
@fredpi Are you ok with merging this PR? I will rebase #102 after merging

Sources/SFSafeSymbols/Symbols/SFSymbol+AllSymbols.swift
```diff
@@ -6,20 +6,31 @@ extension SFSymbol {
     public static var allCases: [SFSymbol] { Array(allSymbols) }
 
     public static var allSymbols: Set<SFSymbol> = {
+        var result = allSymbols1_0
+        if #available(iOS 14.0, tvOS 14.0, watchOS 7.0, *) {
+            result.formUnion(allSymbols2_0)
```
fredpi: [Optional] I would suggest to use another naming: `symbolsAvailableSince2_0`. It's an `internal` variable and one could argue that naming is not so important, but `allSymbols2_0` is not the proper description of what's behind the variable.

fredpi (approved):
@StevenMagdy Thank you very much! I just proposed one optional improvement; you can decide by yourself whether you want to implement it or not and then merge this PR.

Sources/SFSafeSymbols/Symbols/SFSymbol+AllSymbols+1.0.swift
```diff
@@ -1647,7 +1647,7 @@ extension SFSymbol {
             zCircleFill,
             zSquare,
             zSquareFill,
-            zzz
+            zzz,
```
fredpi: Ah, and one other little detail: I would suggest not to put a comma behind the last entry of the set

StevenSorial:
Feedback implemented. Thanks for the review.

PR #102, opened 2022-04-09
Add version 1.1 and 3.3 symbols

StevenSorial:
This adds 3.3 symbols which were released with 15.4. While we can wail for the actual SFSymbols.app, it's been a month and I think it's safe to say that it's not going to be released, similar to what they did when they skipped 2.2.

Adding the symbols based on just info from `name_availability.plist` have a couple of disadvantages:

* Symbol preview is not available as a result of how it's currently generated. I think the current "No preview available" is good enough until the app is released.
* Unavailable layerset info. I've added "Unavailable layerset information" for 3.3 symbols.

This PR also introduces a new version (1.1), fixing some symbols that were not available in iOS 13.0 but actually available in 13.1.

Data taken from macOS 12.3.1: `./System/Library/CoreServices/CoreGlyphs.bundle/Contents/Resources/name_availability.plist`


Sources/SFSafeSymbols/Symbols/SFSymbol+1.1.swift
```diff
@@ -0,0 +1,155 @@
...
+// 1.1 Symbols
+@available(iOS 13.1, macOS 11.0, tvOS 13.0, watchOS 6.1, *)
+public extension SFSymbol {
+    /// 􀞸
+    /// Single Localization, 2 Layersets
+    ///
+    /// Layersets:
+    /// - Monochrome
+    /// - Hierarchical (iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0)
+    static let arrowUturnLeftCircleBadgeEllipsis = SFSymbol(rawValue: "arrow.uturn.left.circle.badge.ellipsis")
```
fredpi, line 12: Have you tested whether these 1.1 symbols are indeed only available on 13.1 and cause a crash when used on iOS 13? That would mean that the official app gives wrong availability information leading to crashes... If they work on iOS 13, there must be a mistake in the plist that you extracted from the OS.

Sources/SFSafeSymbols/Symbols/SFSymbol+3.3.swift
```diff
@@ -0,0 +1,47 @@
...
+    /// Single Localization, ? Layersets
+    ///
+    /// Unavailable layerset information
+    static let personBadgeKey = SFSymbol(rawValue: "person.badge.key")
+
+    /// No preview available
+    /// Single Localization, ? Layersets
+    ///
+    /// Unavailable layerset information
+    static let personBadgeKeyFill = SFSymbol(rawValue: "person.badge.key.fill")
```
fredpi, line 46: Optional: I would rather write `Layerset information unavailable`, but that's a matter of preference, so decide how you want

Sources/SFSafeSymbols/Symbols/SFSymbol+3.3.swift
```diff
@@ -0,0 +1,47 @@
...
+    static let keyViewfinder = SFSymbol(rawValue: "key.viewfinder")
+
+    /// No preview available
+    /// Single Localization, ? Layersets
+    ///
+    /// Unavailable layerset information
+    static let personBadgeKey = SFSymbol(rawValue: "person.badge.key")
+
+    /// No preview available
+    /// Single Localization, ? Layersets
```
fredpi: I'm not very happy with the `? Layersets` notation. I would just leave it out entirely from the short description. If the user is interested in more detailed information (i. e. why the number of layersets isn't specified), they can have a look at the detailed description.

fredpi (changes requested):
@StevenMagdy Thanks for the PR! That's a very interesting discovery!

If we indeed decide to use the file from the OS (which seems to hold more information: 1.1 symbols), it should be documented in the [`CONTRIBUTING.md`](https://github.com/SFSafeSymbols/SFSafeSymbols/blob/stable/CONTRIBUTING.md) where it still says that all files are either obtained according to the comment at the top of the respective file or copied from the Package Contents of the SF Symbols app:

> Some of the files are created manually (there's a comment at the top of these files, explaing how it is done), some other files are copied from the Package Contents of the SF Symbols app.

StevenSorial:
> If we indeed decide to use the file from the OS

I think there is confusion about the OS `name_availability.plist` that I should clarify. This file has always been **identical** to the one inside the app (if the versions match), meaning that the file in macOS 12.1/12.2 is identical to the file inside the `SFSymbols.app` version 3.2, and actually contained incorrect 1.0 symbols. If the latest version of the app is available, we should prefer it just because it's simpler to explain in [`CONTRIBUTING.md`](https://github.com/SFSafeSymbols/SFSafeSymbols/blob/stable/CONTRIBUTING.md).

The discrepancy between the two files happened because the OS file is for the unreleased app version 3.3. This means that apple itself (prior to the latest OS releases) was confused about what symbols were in version 1.0. (i think its because 13.1 was released 5 days after 13.0)

Sources/SFSafeSymbols/Symbols/SFSymbol+1.1.swift
StevenSorial: I added some explanation below about why 1.1 exists in this file and not the app one but to answer your question, yes I tried the 1.1 symbols on a 13.0 simulator and it did crash.

StevenSorial:
Changed wording for unavailable layerset info, and added CHANGELOG entries

Sources/SFSafeSymbols/Symbols/SFSymbol+1.1.swift
fredpi: Thank you! That's very interesting! Bummer we didn't catch that with the CI before, but unfortunately we only test on iOS 13.7, not iOS 13.0

fredpi (approved):
@StevenMagdy Funny how Apple themselves missed the 1.1 version. But if you expect the file specified in the `SF Symbols` app to also include those 1.1 symbols in future updates, that won't pose any problem.

Thanks for the PR, you can merge this if you want. I will then create a release called `3.3`, right?

fredpi:
Ah, one last thing: The table in the README needs to be updated specifying support for 3.3 symbols.

StevenSorial:
Yes, I think the next version should be 3.3.
I will update the README and merge 🎉

PR #106, opened 2022-05-05
Switch runtime localization to be dictionary-based

StevenSorial:
Switched runtime localization to be dictionary-based as described in #98


fredpi:
Thanks @StevenMagdy!!

I will have an in-depth look later, but just wanted to say already that with this change, the `README` documentation must also be changed (and a `CHANGELOG` entry be added)

fredpi (changes requested):
@StevenMagdy Sorry for the long delay reviewing this – but everything looks good to me!

I would just suggest to add a new test that tests for one sample symbol to be defined in `TestHelper.swift`(e. g. `character`) (or for all symbols, but that's probably too much) whether it has the same localizations before and after serialization. While it's obvious that with this implementation, the test will succeed, it would be good to ensure that the statement from the README regarding dynamic symbol localization remains true even with possibly upcoming changes to the implementation.

StevenSorial:
@fredpi ping

fredpi (approved):

fredpi:
@StevenMagdy Sorry for the long delay 🙈 and thanks for the adjustment! Feel free to merge and release a new version (if you have the time – otherwise I may manage to do it next week)

StevenSorial:
@fredpi I don't think we are in a rush for an update. We can wait for the 4.0 release.

PR #110, opened 2022-07-23
Add new image initializers

StevenSorial:
* Add new iOS 16 initializers that support variable images.
* Add a missing UIImage initializer that takes UITraitCollection, introduced in iOS 13.
* Matched Apple docs for the initializers.
* Deprecated initializers for `UIButton`. I think these initializers are the wrong type of convenience since they hide the possible `UIImage.Configuration` and `Double` values for `UIImage`. I think we should only replicate the initializers that take symbol names as `String`. I don't have a strong opinion about this and can revert it if it's not agreed upon.


StevenSorial:
Tests were failing because Bitrise is using Xcode 13.x
Switched Bitrise to use Xcode 14.x and restarted the tests

StevenSorial:
@fredpi The tests are now compiling successfully on Bitrise but the `Install iOS simulators` step is always failing even on `main`. I tried `xcversion simulators` on my machine and it returns an empty list for Xcode 14.0, So it's not related to Bitrise.

I am reverting Bitrise to use Xcode 13 until further investigations.

fredpi:
@StevenMagdy Thanks for the PR!

Unfortunately, the CI is a real mess currently. In The Xcode 14 Beta 5 Release Notes, it said that the simulator issue would be fixed, but installing simulators from terminal is still not working... For now I configured every PR to `stable` to test with Xcode 13 and every PR to `experimental/4.0-support` to test with Xcode 14 (and only iOS 16). So it may be that we don't spot problems directly this way, but `stable` is built every day with both configurations, so we will notice then at the latest.

The PR looks good to me; I would just suggest to also update the README documentation and maybe remove the UIImage test now that it's deprecated. What do you think?

In `experimental/4.0-support` I updated to the latest metadata from Xcode 14 Beta 5 and generally changed the source location for some metadata files to Xcode – the SF Symbols app doesn't get updated regularly and there are quite significant changes, so it will be sensible to always use Xcode as the primary metadata source. Unfortunately, there's no way (at least I don't know one) to obtain the symbol previews from Xcode metadata, so for the new symbols from the latest beta, previews are missing. But we can fix that as soon as the final SF Symbols v4 app with (hopefully) updated metadata is available.

StevenSorial:
@fredpi Removed `UIButton` from the tests and README. Tests were also failing because one of the added initializers is not available for watchOS.

Issue #107, opened 2022-06-06
Support for SF Symbols 4


fredpi:
As can be seen [here](https://developer.apple.com/sf-symbols/), there will be a new version of SF Symbols: **SF Symbols 4**. Currently, the SF Symbols 4 app isn't available to download, but it probably won't last long until we learn all about it.

This issue can be used to collect ideas / track progress on the SF Symbols 4 support.

fredpi:
I updated

- layerset_availability.plist
- name_aliases_strings.txt
- name_availability.plist
- symbol_names.txt
- symbol_previews.txt

from / using the SF Symbols 4 Beta app and regenerated the code. Fortunately, everything went fine, so there's now experimental v4 support: [`experimental/4.0-support`](https://github.com/SFSafeSymbols/SFSafeSymbols/tree/experimental/4.0-support).

I think there are no changes / new features in SF Symbols 4 that require an adaptation of our generator code; but correct me, if I'm wrong.

We just need to update `symbol_restrictions.strings` and `symbol_restrictions_missing.strings`. @StevenMagdy I noticed there's no documentation on how to update them; if I remember correctly you retrieved them from the OS, right? Maybe we can add a file to the `Resources` folder which describes how to update these two files.

fredpi:
Okay, there is one minor issue:

In the `localization_availability.plist`, there is an entry `square.3.layers.3d.down.forward.slash.rtl`, but not an entry for the "base variant" `square.3.layers.3d.down.forward.slash`. In the SF Symbols app, therefore, the symbol is not shown at all. And indeed, it doesn't exist in the OS (causing a crash). Still, SFSafeSymbols assumes that `square.3.layers.3d.down.forward.slash` must exist. Same goes for `square.3.layers.3d.down.backward.slash`.

This can be fixed by making the generator tool ignore localizations if the base variant symbol is not listed in the `localization_availability.plist`. To avoid conflicts, let's wait until #106 is merged before implementing this. Implementing this fix is also a good first issue, if someone is interested.

fredpi:
I pushed a new commit with the new symbol restrictions file from Xcode 14 Beta. Thanks @StevenMagdy for documenting the information where to find it (via #108).

Interestingly, the Xbox and PS symbols introduced in SF Symbols v3 are now also restricted despite not representing Apple products:

`This symbol may not be modified and may only be used to refer to Microsoft’s Xbox.`

[...]

PR #111, opened 2022-09-10
Support for SF Symbols 4.0

fredpi:
This is the PR with support for SF Symbols 4.0, closing #107.

I updated the code with the current metadata files from SF Symbols Beta (Build 78) and Xcode 14 RC. It seems like the metadata files from these sources harmonize with each other, i. e. there are no lacking symbol preview etc.

The PR also includes #110.

As the CI issue discussed in #110 still exists, I have now setup the CI to always test everything against the latest OS. Daily builds are however still done with older OS to ensure backward compatibility.

I suggest to release version 4.0.0 as soon as this PR is merged.


fredpi:
CocoaPods Lint is causing issues again... (I think it's a combined issue of Xcode 14 RC not including support for Monterey and the tvOS and watchOS simulators not being installed by default).

I don't know how to properly handle this... Delay the release until Monterey is out? Delay the release via CocoaPods only? Drop CocoaPods entirely? (after all, installation as a Swift package is now super easy)

StevenSorial:
I have fixed the macOS issue by checking for the `AppIntents` framework which is only available on the latest OS releases. We should remove the check once Xcode 14.1 is released.

I will try to fix the other issue related to the unavailable simulators.

StevenSorial:
Fixed 🎉 

> tvOS and watchOS simulators not being installed by default

This turned out to be false for Bitrise. The output log of the `Install simulators` step starts with `Nothing to download`, which is the result for `xcodebuild -downloadAllPlatforms` command, so I think Bitrise already pre-installs tvOS and watchOS SDKs to save bandwidth.

After replicating the issue on my machine I found out that Cocoapods lint wasn't able to find the `Apple Watch SE (40mm) (2nd generation)` simulator even though it exists, probably because it was not paired to an iPhone simulator.

I deleted all existing simulators and created the needed ones manually. I think this is the better approach going forward to have more control over the workflow and reduce CI failures.

This is the added script to the end of the `Install simulators` step:
```bash
xcrun simctl delete all;
phoneID=$(xcrun simctl create "iPhone 14" com.apple.CoreSimulator.SimDeviceType.iPhone-14);
watchID=$(xcrun simctl create "Apple Watch SE (40mm) (2nd generation)" com.apple.CoreSimulator.SimDeviceType.Apple-Watch-SE-44mm-2nd-generation);
tvID=$(xcrun simctl create "Apple TV" com.apple.CoreSimulator.SimDeviceType.Apple-TV-1080p);
xcrun simctl pair $watchID $phoneID;
```

fredpi:
@StevenMagdy Awesome! You're brilliant!

After running the primary workflow on this branch (which as described earlier unfortunately doesn't run automatically, but only on manual triggers and daily on `stable`), I discovered another issue with the new variable value initializers being unknown to Xcode 13. I fixed it with the same `#if canImport(AppIntents)` check that you already used for fixing the macOS CocoaPods issue.

~Let's see if the CI runs successfully now.~ Update: The CI now runs successfully with both workflows.

Also, let me know whether we can merge this and release after the merge.

StevenSorial (approved):
@fredpi Thanks for the PR. Sorry for taking too long to review.

I'm not sure about the workaround for the two unexpected symbol names, my worry is that it might be an expected pattern and not a bug, in that case, we might miss some symbols. Another workaround might be declaring `let ignoredSymbols: [String] = []` with just the two symbols, that way we will catch it if it's repeated in future versions. We can release the current workaround and rethink it later.

It also might be smarter to make copying/parsing the resource files part of the script since most of them now are in `Xcode.app`. This can be done in a later PR.

fredpi:
@StevenMagdy

> I'm not sure about the workaround for the two unexpected symbol names, my worry is that it might be an expected pattern and not a bug, in that case, we might miss some symbols. Another workaround might be declaring let ignoredSymbols: [String] = [] with just the two symbols, that way we will catch it if it's repeated in future versions. We can release the current workaround and rethink it later.

Hmm, I think there's really no _valid_ use case for this. And if we wanted to support symbol variants where the base symbol is missing, we would need some special syntax like `enum mySymbolWhereTheBaseVariantIsMissing { static let rtf = ... }` which probably isn't worth it given the SF Symbols app also doesn't support these kinds of symbols. But maybe we can add a print statement to monitor which symbols are effected by the fix. Would you like to create a PR?

> It also might be smarter to make copying/parsing the resource files part of the script since most of them now are in Xcode.app. This can be done in a later PR.

I'm a bit torn on this. It certainly makes the process easier, but we might miss something (e. g. new metadata files) if we don't look regularly at the resources folder "in person". But it would still be quite cool to have such a feature. 😎

PR #115, opened 2022-11-25
Remove unneeded SDK workarounds

StevenSorial:
The SDK workarounds are now not needed in Xcode 14.1

### DIFF
--- a/Sources/SFSafeSymbols/Initializers/AppKit/NSImageExtension.swift
+++ b/Sources/SFSafeSymbols/Initializers/AppKit/NSImageExtension.swift
@@ -13,8 +13,6 @@ public extension NSImage {
         self.init(systemSymbolName: systemSymbol.rawValue, accessibilityDescription: description)!
     }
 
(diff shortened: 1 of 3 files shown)


fredpi:
@StevenMagdy I'm wondering whether it's actually good to remove these workarounds – when someone is using the upcoming new SFSafeSymbols version with an older version of Xcode (which is definitely not a discouraged use case), I think problems would occur if these checks would be removed. Or am I getting it wrong?

StevenSorial:
Yes, I think people with any Xcode less than 14.1 will have compile errors with the next version (I'm not 100% sure). But I don't believe that is something new with Xcode 14.1 or 14.0, SFSafeSymbols has had an implicit requirement of Xcode 11.0 since day 1.

If you think that is too quick to force Xcode 14.1, We can wait until 14.2 and then merge this PR.

fredpi:
@StevenSorial As this is somehow a breaking change (removing compatibility with Xcode 14.0 which is still in use by many) I'd suggest to postpone this until the next major release, okay?

StevenSorial:
Ok, let's keep it open until then.

PR #119, opened 2023-01-27
Allow optional symbol for SwiftUI's `Label`

StevenSorial:
When using SwiftUI, i find myself writing `Label("My Label", systemImage: showImage ? "plus" : "")`, which doesn't translate well when using `SFSafeSymbols`.
Option 1:
```
If showImage {
	Label("My Label", systemSymbol: .plus)
} else {
	Text("My Label")
}
```

Options 2:
```
Label("My Label", systemSymbol: showImage ? .plus : .init(rawValue: ""))
```

Option 3:
```
Label("My Label", systemSymbol: .plus)
	.labelStyle(showImage ? .titleAndIcon : .titleOnly)
```
I think option 1 is too verbose, option 2 requires the user to understand how the SFSymbol initializers work, option 3 I think Is the most acceptable but I think it's a little weird to always assign an image.

This PR makes the symbol optional to be:

```
Label("My Label", systemSymbol: showImage ? .plus : nil)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 
 ### Changed
 
-- None
+- Changed SwiftUI's `Label` initializer to take an optional `SFSymbol` (By [Steven Sorial](https://github.com/StevenSorial))
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


fredpi (changes requested):
@StevenSorial Thanks, LGTM!

Just two small details:

- Could you add a changelog entry?
- Could you document the behavior when a `nil` value is passed (in the documentation of the function)?

fredpi (approved):

StevenSorial:
@fredpi  Do you mind if we released 4.1.1 with this change?

fredpi:
@StevenSorial That'd be good 👍 Feel free to do the release :)

PR #121, opened 2023-06-21
Add missing initializers for SwiftUI

david-swift:
Hello,

Thanks to all of you for maintaining such a great project! 

There seem to be initializers for two SwiftUI views: 
- `Image`
- `Label`

Those are definitely the most important views for displaying symbols, but there are two other SwiftUI types with initializers containing a `systemName: String` parameter: 
- [`MenuBarExtra`](https://developer.apple.com/documentation/swiftui/menubarextra/)
- [`ContentUnavailableView` (Beta)](https://developer.apple.com/documentation/swiftui/contentunavailableview/)

It would be nice to have "safe" initializers for all the SwiftUI views supporting SFSymbols.

I added a "safe" version for all the initializers provided by Apple for `MenuBarExtra` and `ContentUnavailableView` with the `systemName: String` parameter, as well as test cases for both of the types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 
 - Add support for SF Symbols 6 (By [Rod Brown](https://github.com/RodBrown1988))
 - Added initializers for SwiftUI Button.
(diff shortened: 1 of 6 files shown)


guillaumealgis:
I haven't looked to much at the code here, but it seems like a good idea to cover all cases of `systemName: String` arguments in the system frameworks. Is there something blocking this PR? How could one help get this merged?

Thanks!

StevenSorial:
@david-swift can you please rebase your branch so I can take a look

PR #122, opened 2023-09-26
WIP: Added symbols as fallback for previous iOS versions

philprime:
Hi there,

first of all, thank you for this open source project! 
We truly appreciate you working on making it safer to work with SFSymbols.

Due to many SFSymbols only being available in newer versions, it becomes hard to have consistent symbols over multiple OS versions. In some cases there are good alternatives, but in others there simply isn't. 

Luckily we noticed that is possible to ship SFSymbols exported from the official SFSymbols.app by Apple and include them in a `.xcassets` asset catalog.
We went ahead and extended SFSafeSymbols to have the same workflow as we do with symbols already available.

This gives us the following way of handling different versions:

```swift
if #available(iOS 16.0, *) {
    return SFSymbol.beachUmbrella
} else {
    return SFSymbol.beachUmbrellaFallback
}
```

Now I wanted to open up this pull request to discuss if you are interested in extending SFSafeSymbols with our fallback approach (hence the WIP in the title). 
If you do, I am also committed to go through multiple iterations together to get this improved.

**Some technical details:**

- The asset catalog needs to be listed in the Package resources. As this increases the package size, it might be a better approach in moving it to another target in this Swift package, i.e. `SFSafeSymbolFallback` or similar
- Symbols can be added from the SFSymbols.app by selecting the symbol, then **File > Export Symbol > Export** and then dragging it into the Asset Catalog in Xcode. Maybe this can be combined with the SFSafeSymbols generator.
- As we are using [SwiftGen](https://github.com/SwiftGen/SwiftGen) in many parts of our apps, we decided to generate the enum based on the asset catalog using it. If you prefer a different approach without this dependency, we could maybe add this to the SFSafeSymbols generator as well.


philprime:
Any feedback on this proposal?

StevenSorial:
sorry for the delay. I will get to it in the coming days.

StevenSorial:
Thanks for the nice PR and the explanation. I see this as an app-side feature and shouldn't be included in the library:

* I noticed that you only included the fallback for just the symbols introduced in iOS 16.0 (v4). for us to add this feature, we would have to add a fallback for all the versions.  basically all the symbols except v1 symbols.
* As you mentioned, this would increase the library size quite significantly. The library is already very large, and I'm already thinking how to make it smaller.
* This feature doesn't use any internal code. so the initializers and the assets can be included the app, and it would be used the same way as if this PR was merged.

That being said, i really appreciate the effort that was put into this PR.
Thanks.

philprime:
Sorry for my delay on my side.

> I noticed that you only included the fallback for just the symbols introduced in iOS 16.0 (v4). for us to add this feature, we would have to add a fallback for all the versions. basically all the symbols except v1 symbols.

I believe this could also be automated with a generator, as you already do for the symbols themselves, and maybe build packages per iOS version. Not entirely sure if it is possible tough.

> As you mentioned, this would increase the library size quite significantly. The library is already very large, and I'm already thinking how to make it smaller.

I agree, and it's probably also the biggest downside of shipping all the missing symbols. Maybe there would be a way of make it more dynamic, but that probably also exceeds the scope.


In that sense I agree that there is probably no point in this PR anymore, therefore feel free to close it.

PR #128, opened 2024-02-06
Swift 6 consideration: make SFSymbol Sendable

Dean151:
## Motivation
Swift 6 is coming (one day?) and Strict Concurrency Checking is the new safe default to build new projects.
Since latest Xcode beta (15.3 beta 2), Strict Concurrency warnings are becoming more powerful … and trigger warnings on SFSymbols and related not being sendable.

## An easy fix
Therefore, the fix is easy. A class is sendable when it's immutable and not overridable, or when it manage its internal state safely. In our case, we don't have the "not overridable", but it's manageable! The class only store immutable data (rawValue) and it's not overridable outside of the module, because it's not open. Therefore, we can safely assume it's Sendable by default, and add `@unchecked Sendable`.
Unchecked is a little bit misnamed, maybe? It's not that we haven't checked as a developer. It means the compiler hadn't checked it. 
And like we say. With great power comes great responsibility.


StevenSorial:
Thanks for the PR. can you please rebase before I can take a look?

Dean151:
Hi @StevenSorial !
The rebase is complete !
Have a nice review :)

Sources/SFSafeSymbols/Symbols/SymbolLocalizations.swift
```diff
@@ -1,6 +1,6 @@
 // Don't touch this manually, this code is generated by the SymbolsGenerator helper tool
 
-public enum Localization: String, Equatable {
+public enum Localization: String, Equatable, Sendable {
```
StevenSorial: Correct me if I'm wrong, but I don't think simple enums need an explicit conformance.
if that's the case then we can remove it

StevenSorial:
Thanks. I left one comment.
Could you also please add a changelog entry with that change?

Dean151:
@StevenSorial I can't see your comment yet, is it still "pending for sending" in the review feature of Github?

StevenSorial (changes requested):

Sources/SFSafeSymbols/Symbols/SymbolLocalizations.swift
Dean151: I agree. Since String is Sendable, so is the enum, implicitly. Removing it right away!

Dean151: And it's done :)

StevenSorial:
Thanks for the changes. Feel free to credit yourself in the changelog using the same format as the rest.
I will merge the PR tomorrow.

